PR #7, opened 2016-01-11
Added information to the page for paintutils files

lxndio:
Added basic information on the file format itself and the technological information.

### DIFF
--- a/File Formats/Images/Paintutils Image Format.md
+++ b/File Formats/Images/Paintutils Image Format.md
@@ -2,11 +2,12 @@
 
 ## Quick information
 
------------ | ------------------------
-Type		| Image file format
-MIME		| image/paint
-Extensions 	| .nfp
-Used by 	| paint, paintutils (API)
+| Information |                         |
+| ----------- | ----------------------- |
+| Type        | Image file format       |
+| MIME        | image/paint             |
+| Extensions  | .nfp                    |
+| Used by     | paint, paintutils (API) |
 
 ## Technical details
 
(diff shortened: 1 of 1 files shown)


demhydraz:
:+1: leaving that placeholder there is quite ugly. (@oeed ಠ_ಠ)

SquidDev:
Would it be sensible to use code blocks (``...``) around the file extensions and mime types? If you look in [Wikipedia's page](https://en.wikipedia.org/wiki/Media_type) is uses them and personally I think it look nicer.

oeed:
Nice job! Yeah I basically had the placeholder there just so people could get a basic idea of the folder structure.

I'd also agree with SquidDev about the code blocks.

One thing that looked like was missing was also the space character represents transparency. I could be wrong, but based on some of the CraftOS 2 videos it appears as if Dan has given these files the extension 'image'. It might be another format though actually, so it might be worth holding off on that for now.

lxndio:
I will change the mentioned things today. Should look good then.

oeed:
@lxndio I'd also note what was mentioned in #10. Essentially the 'used by' might cause issues later on.

lxndio:
@oeed Yeah, I first thought it would make sense to name some examples but as you say, it can cause quite some issues later when others are trying to fill in their programs etc.

oeed:
Last two nit picks, it might be worth adding transparent to the colour table.

Also, the example image has a missing pixel at the bottom right.

lxndio:
@oeed I thought, it would be fine to mention transparency at pixel data because transparent isn't a color.
The missing pixel was thought to show that not every line has to contain the same amount of pixels, probably that should be mentioned down below then. 
Let me know what you think about that, please.

oeed:
Ah, yeah okay that's understandable.

PR #8, opened 2016-01-11
Executable file standard

Admicos:
I think this should be standartised since that would help the developers (detecting what is an executabe and what isn't, what do they do etc.) and users (probably the same as developers)

I am open for ideas and possible additions - deletions for this.

### DIFF
--- a/File Formats/Executables/Executable Format.md
+++ b/File Formats/Executables/Executable Format.md
@@ -5,8 +5,9 @@
 Executable files are the programs the user or another program executes.
 
 Their extension should be ".mc" and the first two lines of the executable code should be comments about what it is.
-  
-**Example:** ```Lua
+
+**Example:**
+```Lua
 -- Example code by Admicos
 -- This program is an example for the Executable Format standard.
 
--- a/File Formats/Executables/Executable Format.md
+++ b/File Formats/Executables/Executable Format.md
@@ -1,0 +1,15 @@
+# Executable Format
+#### MIME: lua/executable
+#### Extension: .mc
+
+Executable files are the programs the user or another program executes.
+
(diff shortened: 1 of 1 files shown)


demhydraz:
:-1: this is quite useless. Here's a way to detect what is an executable and what isn't:

```lua
return ({loadfile(file)})[1] ~= nil
```

viluon:
Before publishing a standard proposal, ask yourself these questions:
- [ ] **Does your proposal/format really need to be standardized?** Standards exist to increase cross-platform compatibility and keep reinventing the wheel to a minimum. In many cases, however, the proposal can be overly specific or so simple that standardizing it yields little to no benefits.
- [ ] **If your standard gets accepted, will its users benefit from it?** Does your proposal help solve frequent issues related to the category it should be used in? Does it help with compatibility (esp. backward and forward compatibility with future standard revisions)? Is it easy to use? 
- [ ] **Is your proposed standard easy to use?** Does it require third-party code or brand new code to be written in order to be used properly, or do you pack needed utilities with the proposal? Are these utilities efficient? 

No, no and no again.

In this case, @Admicos, I see these specific problems with your proposition:
- This standard **doesn't help anything**
- You have provided no utilities for parsing the format
- Most CC source files to date use the `.lua` extension or don't use one at all
- This standard doesn't even ensure that the file _is_ in fact executable, and could easily contain compile errors

demhydraz:
Danke.

oeed:
I'd have to agree. Especially now that CraftOS 2 appears to use file extensions far more heavily based on some of the videos, so .lua would probably be the accepted standard. As such I think we need to create a standard which mentions using .lua.

PR #10, opened 2016-01-11
Add Standard Proposal Guidelines

### LINKED ISSUE #9: Standard Proposal Guidelines
@SquidDev's comment on #7 reminded me that there are no guidelines on how a Markdown proposal document should look like. Adding one should be fairly easy!

viluon:
Closes #9. Still in progress, comments are very welcome!

### DIFF
--- a/Standard Proposal Guidelines.md
+++ b/Standard Proposal Guidelines.md
@@ -5,13 +5,13 @@
 ---
 
 # _Standard Name_
-| Information	|																						|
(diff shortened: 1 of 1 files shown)


oeed:
I'd agree with @SquidDev about the 'used by', it's going to get messy and hard to maintain. Otherwise it looks good (although we'll see if any other changes might be useful).

oeed:
As SquidDev mentioned in #7, it'd be better to put the MIME and extension in code blocks.

Issue #6, opened 2016-01-11
Official Recommendations


viluon:
Maintain a list of preferred (recommended) formats for different categories. Multiple formats for the same category can be recommended. For a standard to be recommended, it has to be accepted and proven to be used by a large enough number of users (this might be very hard to determine, but I'll leave this up on the contributors and owners of this repository :stuck_out_tongue:). A good example of a standard to be recommended is [NFP](https://github.com/oeed/CraftOS-Standards/blob/master/File%20Formats/Images/Paintutils%20Image%20Format.md), the default image format for `paint` files.

oeed:
Yeah that is a good point. Maybe mentioning it in the category README. I'd say we should only be accepting multiple standards for one thing (i.e. image formats) if they have different use cases or they're all highly used.

viluon:
_Bump_, recommendations need an explanation on what they are. A list of recommended standards can already be viewed [here](https://github.com/oeed/CraftOS-Standards/pulls?utf8=%E2%9C%93&q=label%3Arecommended) (although it was me who just labeled the recommended standards, which were listed in the [Image README](https://github.com/oeed/CraftOS-Standards/blob/master/File%20Formats/Images/README.md), and it _wasn't me_ who put them there, so please keep these lists **synced**, it's very important), but we should create a file about official recommendations maybe in the `/Tutorial/` folder (currently only in the `tutorial` branch). The rules for accepting a standard as `recommended` should also be outlined more clearly.

PR #11, opened 2016-01-11
Created image file format README with official recommendations

oeed:
The main issue involves the official recommendations #6 

I personally do not think the paint image format should be a recommended format. Although it has not yet been publicly been released, UCG (by ardera) is a far more efficient format (honestly, it can shrink file sizes to below 20% of the original size).

The issue is, however, unless Dan implements UCG in the paint program (which he should IMO) people are going to use paint, especially until there's an editor available for UCG images.

As UCG is as efficient in lower sizes, perhaps we only recommend paint for small images?

### DIFF
--- a/File Formats/Images/README.md
+++ b/File Formats/Images/README.md
@@ -1,0 +1,13 @@
+# Image File Formats
+
+## Official Recommendations
+
+Rather than creating your own format or using a lesser used format, it is recommended to use the following formats based on your needs.
+
+**Images that require both text and colour in each pixel (advanced computers)**: [NitroFingers Text](todo)
+
+**Images that only require 1 colour per pixel**: [Universal Compressed Graphics Format](todo)
+
+## Proposal Requirements
+
+Image file format proposals must follow the [File Format proposal requirements](https://github.com/oeed/CraftOS-Standards/blob/master/File%20Formats/README.md#proposal-requirements).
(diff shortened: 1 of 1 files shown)


viluon:
One way I thought of the official recommendation idea #6 was that this repository and its maintainers would be regarded as an authority by the rest of the community. Saying that UCG is the recommended image format would be equal to saying that it should be implemented in every painting program (which is my honest opinion) and every program that does not follow the recommendations should be essentially fixed to do so. As soon as UCG is standardized, I would recommend it.

oeed:
The question is though, do we flat out not recommend paint at all? The problem with UCG is the code is actually pretty complicated. It uses Huffman trees and reading & writing both take a few hundred lines rather than the single function call that's built in to paintutils.

Ideally @dan200 woould change paintutils to use the UCG format, but we can't guarantee that at all.

viluon:
If there was an official format for data transmission in CC that you knew was flawed, would you recommend it? If UCG is superior, there is no point in recommending NFP _instead_. Both can be recommended, but I would prefer UCG. Don't forget that the standard proposal should already contain utilities and example code to deal with complex algorithms, and users can really just copy - paste that.

oeed:
In that case yeah, I'd say we shouldn't _recommend_ the format.

PR #12, opened 2016-01-12
Nitrogen Fingers Text Image Format

oeed:
This format is an existing format, so checking of the facts is more important rather than actually approving the topic.

### DIFF
--- a/File Formats/Images/NitroFingers Text.md
+++ b/File Formats/Images/NitroFingers Text.md
@@ -1,0 +1,58 @@
+# NitroFingers Text
+
+## Quick information
+
+| Information |                           |
+| ----------- | ------------------------- |
+| Type        | Image file format         |
+| MIME        | `image/nft`             |
+| Extensions  | `.nft`                    |
+
+## Technical details
+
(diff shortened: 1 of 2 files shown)


oeed:
@viluon whilst I too place space between brackets, I'm not really sure whether that styling should be used here as it's not overly common.

@SquidDev Yeah you're probably right there. If people don't understand how to do that they'll probably just use the premade API anyway.

viluon:
@oeed I don't enforce the additional whitespace, but your code used different formatting at different places..

oeed:
@viluon right.. so I did actually :stuck_out_tongue:

demhydraz:
:+1: from me.

viluon:
_bump_ can we accept this @oeed?

oeed:
@viluon I'm still at her house and only have my phone, so I can't edit it right now, but I think we should actually move the character/byte section to teminology or something similar rather than having it in here. I might be home in a day or two, but if you are able to do it now we can merge it.

viluon:
I'll unpack my backpack from the mountain trip and get right to it @oeed

oeed:
Awesome!

Issue #22, opened 2016-01-13
Standardize Hidden Files


oeed:
Standardising the files that should be ignored/hidden might also be useful.

For example, all files with the '.' prefix are hidden/ignored as well as things like _Desktop.ini_ (we'd need a standard list).

lxndio:
Starting to work on this.

lxndio:
Can be closed. See #27.

viluon:
@lxndio use "Closes X" in your PR to close this issue automagically after merging.

viluon:
I won't label this as accepted since #27 wasn't closed/merged yet.

oeed:
I see where you are coming from, although the concept has been accepted, the PR and issue are two separate things.

It's your decision though, I'm happy with either approach as long as we're consistent :smile:

viluon:
#27 was now closed, there's no need to standardize hidden files (separately from an OS standard family) (see my comment on #27). Feel free to reopen if you want to discuss this in greater detail. Labelling as 'rejected' for now.

Also, please don't close issues that are supposed to be solved by a PR @oeed, just use 'Closes #X' in the PR and it'll close automagically after merging, or stay open otherwise (which is what we want!).

PR #23, opened 2016-01-13
Create .gitignore

demhydraz:
- Ignore some OS-created files (seriously, WHO DOES THIS)
- Ignore all files starting with a period.


viluon:
Does not close #22, but accepted

PR #25, opened 2016-01-13
Add Plain Text Format

oeed:
I feel like the description could possibly be longer, and that RFC might be wrong, but I mean... what else can you add?


demhydraz:
:+1:

oeed:
> okay, whatever
> @viluon 

Agreement if I've ever seen one, accepted.

PR #26, opened 2016-01-13
Add Lua Source Code Format

### LINKED ISSUE #16: Lua Source Code Format
_Whilst this is a fairly well known format it is still important that these details are standardised._

Extensions: `.lua` (and any other extensions you know of)
MIME: `text/lua`

oeed:
Closes #16

### DIFF
--- a/File Formats/Code/Lua.md
+++ b/File Formats/Code/Lua.md
@@ -1,0 +1,15 @@
+# Lua
+
+## Quick information
+
+| Information |                           |
+| ----------- | ------------------------- |
+| Type        | Plain text file format    |
+| MIME        | `text/lua`                |
+| Extensions  | `.lua`                    |
+
+## Technical details
+
+*This format is based upon [plain text](https://github.com/oeed/CraftOS-Standards/blob/master/File%20Formats/Documents/Plain%20Text.md) and as such technical details that apply to plain text also apply to this format.*
+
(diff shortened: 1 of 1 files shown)


demhydraz:
:+1:

lyqyd:
There's lots of things in CC that a valid lua file could be:

lua/data
lua/program
lua/library

So why are we collapsing all of them to one mime type?  We are missing an opportunity here.

oeed:
That's a fair point @lyqyd, but how do you specify which one it is? If the system determines the MIME based on the file extension they'll all default to `lua/program`. I guess one solution to this would to be API/libraries a separate extension like `.lib` or something like that.

lyqyd:
Yes, separate file extensions are one way to go.  You have a good point though, that there needs to be a way to tell the metadata system what variety/usage of Lua a file contains.  Is there already a suggestion/draft for a generalized metadata attribute setting standard?

oeed:
@lyqyd I've been working on a proposal [here](https://github.com/oeed/CraftOS-Standards/blob/oeed-patch-metadata/File%20Formats/File%20Metadata.md), it's not quite complete, but the basics are there.

I'd originally proposed a system more alike your RFC, although everyone in Gitter was quite against having a metadata file for each file. Another issue I found with your system was as Silica has a few hundred files metadata started taking up a significant amount of storage space, this format is designed to be as tight as possible.

I don't think it's mentioned yet, but in Silica I currently have it 'guessing' the MIME based on a file's extension; I'm not really sure of another reliable method.

PR #27, opened 2016-01-13
Added information about hidden files and folders

### LINKED ISSUE #22: Standardize Hidden Files

lxndio:
Closes #22


demhydraz:
Squash your commits pls

lxndio:
@demhydraz I did after pushing this. Excuse me, please.

demhydraz:
doesn't look very squashed from here

lxndio:
Closes #22

oeed:
It might be worth holding off accepting this until we decide upon a metadata format, as that file will also need to be hidden.

viluon:
Review:
- **Poorly written**
  - Uses informal language of a less-than-academic level (contracted forms, "you _must_", etc)
  - Does not follow the [Standard Proposal Guidelines](https://github.com/oeed/CraftOS-Standards/blob/master/Standard%20Proposal%20Guidelines.md)
- **Poorly designed**
  - Most operating systems only use metadata to determine whether a file or a directory should be interpreted as hidden, or use the specified leading-dot naming convention in tandem with metadata
  - A file of the name `desktop.ini` or `Thumbs.db` is not necessarily hidden, keeping a list of '_usually hidden files_' is pointless
- **Unnecessary**
  - There is no real need to standardize which files should be interpreted as hidden, as long as such a standard isn't a part of a larger (operating system) standard or standard family, such as [POSIX](https://en.wikipedia.org/wiki/POSIX)

PR #28, opened 2016-01-13
Add folder

### LINKED ISSUE #17: Folder/Directory Standard
_Whilst this isn't technically a file type it is still important that the MIME is standardised._

MIME: `folder`

The reasoning for giving folders a MIME is that some metadata systems allow folders to act like files and as such need a MIME.

oeed:
Might need some technical details, but who knows what they might be.

Closes #17

### DIFF
--- a/File Formats/Folder.md
+++ b/File Formats/Folder.md
@@ -1,0 +1,15 @@
+# Directory
+
+## Quick information
+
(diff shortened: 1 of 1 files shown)


oeed:
No objections; approved.

PR #35, opened 2016-01-20
Miscellaneous Standards

viluon:
A new folder for misc stuff. Also, I think it'd be wise to put repo-administration-related stuff into a separate meta folder (I'll do it in `tutorial`... some time later).

PR #36, opened 2016-01-20
Terminology standard

demhydraz:
Should I put anything here?

I guess I should say that this PR incorporates #35.


oeed:
@demhydraz I think you need to link to, or explain somewhere, things like 'IEEE 754'.

I also don't really think that's the right place for it.

demhydraz:
well blame @viluon

oeed:
The isn't a standard though (is it?) It's more for terminology used in COSs.

demhydraz:
Well, it's _standard_ terminology.

Not all IETF RFCs are standards; See ["Etimology of Foo"](https://www.ietf.org/rfc/rfc3092.txt) and ["        Key words for use in RFCs to Indicate Requirement Levels"](https://www.ietf.org/rfc/rfc2119.txt)

oeed:
Accepted.

ardera:
Why is the word "byte" not allowed? Everyone knows what a byte is, the term "octet" is used far less.

And what about adding things like "uword" = unsigned 2 octet integer and "word" = signed 2 octet integer?

oeed:
Yes you're quite right @ardera. I don't see what's wrong with using byte and you need to explain little and big endian.

Sorry, this was quite a mistake on my behalf letting it through wi such little checking.

Anyone is welcome to make a PR with these changes.

demhydraz:
Because byte is ambiguous. It's not well defiend. I could very well make a processor with 4-bit bytes.

ardera:
then define it as a 8bit binary number

viluon:
Agreed with @ardera.

demhydraz:
It's still ambiguous. Octet has 8 in the name.

ardera:
why is it ambigous if you give it the same definition an octet's got?

oeed:
@demhydraz you're just being pedantic. Just specify that a byte is 8-bits, as 99% of sane people use.

superaxander:
This may be intended but right now the terminology standard states that a signed word can get up to higher values than the unsigned variant. This seems strange to me.

oeed:
I think we need to create a new PR with all these fixes. Anyone, other than demhydraz, who is up to it, please go ahead and create one (from master, not here) with the changes mentioned.

demhydraz:
> other than demhydraz

wow rood

demhydraz:
> This may be intended but right now the terminology standard states that a signed word can get up to higher values than the unsigned variant. This seems strange to me.


Uh, no, you're just reading wrong.

ardera:
actually superaxander is really just reading it wrong

superaxander:
Ok I indeed read it wrong. But may I suggest you use the same number format on each number
So change this line:
```
A word is comprised of two octets (16 bits), ranging from -32,768 to 36,768 when signed or from 0 to 65535.
```
To be:
```
A word is comprised of two octets (16 bits), ranging from -32,768 to 36,768 when signed of from 0 to 65,535 whenwhen unsigned.
```
I think that is what put me off track.

PR #37, opened 2016-01-22
Fix a formatting issue in Terminology.md

demhydraz:
nice reviewing @oeed

### DIFF
--- a/Terminology.md
+++ b/Terminology.md
@@ -20,7 +20,7 @@
 	2.3 **Floats**
 3. **General**  
 	3.1 **Executable**  
-	3.2 **Library**
+	3.2 **Library**  
 	3.3 **Resource**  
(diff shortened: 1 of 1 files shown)


oeed:
This deserves an appropriate label.

demhydraz:
... *sigh* just merge it

oeed:
So no more cock-ups in there?

demhydraz:
Nope.

viluon:
Ehm, you let @demhydraz basically self-approve this @oeed. While in this case it's absolutely fine, this shouldn't happen in the future :confused:

oeed:
@viluon it was just a brainfart we both overlooked and noticed, but the original PR certainly was, well, a cockup that I shouldn't let through.

PR #13, opened 2016-01-12
How to Propose a Standard (Tutorial)

viluon:
A tutorial to get you started (currently work in progress). I am planning on moving tutorial-like stuff from the README into the new tutorial directory. What do you think?

PR #38, opened 2016-01-24
Temporary Directory Standard

ghost:
This is my proposal for a standard path/directory for temporary files. Feel free to take a look and let me know what you think.

I know that it isn't even code, but it's something that should be standardized.

Let me know what you think.

### DIFF
--- a/Miscellaneous-Standards/Temporary Directory.md
+++ b/Miscellaneous-Standards/Temporary Directory.md
@@ -1,0 +1,46 @@
+# _Temporary Directory_
+| Information	|																						                                            |
+|-------------|---------------------------------------------------------------------------------------|
+| Version 		|1.0.0											                        |	
+| Type		  	| Path / Directory																	                                      |
+| MIME		  	| `path/temp`																			                                      |
+| Location	| Temporary Directory is located in `/.temp`	|
+| Deprecates	| No standards have been made for this yet.				                  			|
+
+### Technical Details
+A standard temp directory exists at `/.temp`
+
+This directory is used for storing files temporarily. The point of this directory is to be able to have a standard place to store "junk" files, where the program doesn't have to worry about accidentally overriding another file created by the user. It is NOT a place to store any files that need to be accurately recalled later on.
+
+**Proper Usage in Supporting Programs**
+
+To use it properly, a program can (through normal non-interrupted execution) store files in the `/.temp` directory while it is running, and then should promptly delete the file from this directory as soon as possible.
+
+**Proper Support in Supporting OSes (and other systems)**
+
+The `/.temp` folder should be deleted then recreated at the startup of the system. Access to this directory should be open and unprotected.
+
+### Examples
+
+#### Usage (In a Program)
+```Lua
+local junkFile = fs.open("/.temp/blah", "w")
+junkFile.write("So since I can write here, I'm not at risk of messing a user's stuff up.")
+junkFile.close()
+
+sleep(0)
+
+if fs.exists("/.temp/blah") then
+	local junkFile = fs.open("/.temp/blah", "r")
+	-- junkFile could be nil, so you will want to check for that. There is no guarentee the file still exists - the temp directory is not persistent.
+    local var = junkFile.readAll()
+end
+```
+
+#### Usage (In an Operating System or Other Booting System)
+```
+-- This is a startup file. Append this code to the top of your startup file (or close to it) to adhere to the standard.
+
+fs.delete("/.temp")
+fs.makeDir("/.temp")
+```
--- a/Miscellaneous-Standards/Temporary Directory.md
+++ b/Miscellaneous-Standards/Temporary Directory.md
@@ -2,33 +2,31 @@
 | Information	|																						                                            |
 |-------------|---------------------------------------------------------------------------------------|
 | Type		  	| Path / Directory									
-| Location	| Temporary Directory is located in `/.tmp`	|
+| Location	| Temporary Directory is located in `/tmp`	|
 
 ### Technical Details
-A standard temp directory exists at `/.tmp`
+A standard temp directory exists at `/tmp`
 
 This directory is used for storing files temporarily. The point of this directory is to be able to have a standard place to store "junk" files, where the program doesn't have to worry about accidentally overriding another file created by the user. It is NOT a place to store any files that need to be accurately recalled later on.
 
 **Proper Usage in Supporting Programs**
 
-To use it properly, a program can (through normal non-interrupted execution) store files in the `/.tmp` directory while it is running, and then should promptly delete the file from this directory as soon as possible.
+To use it properly, a program can (through normal non-interrupted execution) store files in the `/tmp` directory while it is running, and then should promptly delete the file from this directory as soon as possible.
 
 **Proper Support in Supporting OSes (and other systems)**
 
-The `/.tmp` folder should be deleted then recreated at the startup of the system. Access to this directory should be open and unprotected.
+The `/tmp` folder should be deleted then recreated at the startup of the system. Access to this directory should be open and unprotected. It can however be hidden by the operating system, however, this is at the discretion of the OS creator (and hopefully a user-modifiable config setting as well). **Ensure that you are only hiding /tmp and not other folders called tmp**
 
 ### Examples
 
 #### Usage (In a Program)
 ```Lua
-local junkFile = fs.open("/.tmp/blah", "w")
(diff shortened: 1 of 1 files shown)


lyqyd:
If this is going to be a thing, it belongs at `/tmp`.

oeed:
I can't really remember ever needing this, but I certainly agree that standardising it is a good idea though.

@viluon If it's not already mentioned in the contributing file, can we make the 'Depreciates' field to only be used when it is depreciating something? @cyanisaac You can remove that row from the table. The same applies for the version, that's really only needed for file formats in which the version is important. The COS number given to this file (which we need to mention in the contributing file too @viluon; they shouldn't add one when they first make a proposal, we will assign one when it has been approved and they can add it) will do the job.

I'd argue that that MIME mightn't be the most appropriate, it should either be based on the directory MIME or just use the directory one.

I feel like it is also worth mentioning that an OS that sandboxes programs should clear out the program's temp folder, rather than the root one.

Also, small nit picks: I don't think the sleep call should be in the example, you're not closing the second file handle, you're not wrapping the file handle calls in if statements and the last code block isn't using Lua.

@lyqyd I'd disagree somewhat, I think at the very least it should be hidden.

demhydraz:
I think it belongs at `/tmp`. That's where I mount my ramfs. That or /dev/shm.

And I think you should provide an API to generate a pseudorandom name (with prefixes, maybe like `appname-XXYYZZ` where X, Y and Z are all distinct random characters) to avoid name clashes.

viluon:
agreed with @demhydraz. Also, could you please comment that over at #13 @oeed? Otherwise I'll probably forget about it :confused:

demhydraz:
> I feel like it is also worth mentioning that an OS that sandboxes programs should clear out the program's temp folder, rather than the root one.

Why tho? Just have a global one and be done with it(™)

ghost:
/tmp is a better place. Also ok I'll remove the unnecessary stuff from the spec - I didn't know what to have and what I didn't need, and I didn't want to submit a terrible request.

Oh, and the sleep(0) is there as an example - when a sleep is called it yields the coroutine, meaning stuff can happen to files during that time. I probably should have explained that in the example, although that's probably not really necessary.

EDIT: And I wrote that example at 11:20PM with myself falling half asleep on the keyboard, hence the other various mistakes.

EDIT 2: Wait, shouldn't it be at /.tmp? I feel like the user shouldn't need to see the temp directory really.

demhydraz:
> EDIT 2: Wait, shouldn't it be at /.tmp? I feel like the user shouldn't need to see the temp directory really.

Then change ls to hide /tmp.

oeed:
> And I think you should provide an API to generate a pseudorandom name (with prefixes, maybe like `appname-XXYYZZ` where X, Y and Z are all distinct random characters) to avoid name clashes.

That's super complicated for something that doesn't need to be, a sandbox does that job for you.

It seems to be that the only reason you guys want tmp is because it's what Linux uses which was probably fairly arbitrary. Complicating it with changing the API seems stupid when you could just prefix a full stop.

Regarding the sleep @cyanisaac I completely understand where you're coming from, but I can just see it confusing some people.

demhydraz:
> It seems to be that the only reason you guys want tmp is because it's what Linux uses which was probably fairly arbitrary. Complicating it with changing the API seems stupid when you could just prefix a full stop.

Linux, FreeBSD, OS X.. basically every sensible OS on the face of the earth.

demhydraz:
Also..
```lua
function fs.openTempFile (appname, mode)
	if not mode and appname:match "[rw]b?" then
		mode = appname
		appname = tostring(os.time())
	end

	local function populateTemplate (template)
		return string.gsub(template, '[xy]', function (c)
			local v = (c == 'x') and math.random(0, 0xf) or math.random(8, 0xb)
			return string.format('%x', v)
		end)
	end

	return fs.open(("/tmp/%s-%s"):format(appname, populateTemplate "xyxy.yyxx"), mode or 'w')
end
```

oeed:
i.e. Some stoned guy in the seventies randomly thought it up. 

I get what you're saying, but to simply follow tradition when it doesn't serve our purpose well is just backwards.

demhydraz:
would you rather have `/Temporary\ Files`?
`/tmp` is short and concise, plus everyone knows what it means.

oeed:
A full stop added doesn't really make it any less concise.

lyqyd:
There's no good reason for it to be hidden.  Just because the user "doesn't _need_" to see it isn't enough to justify hiding it.

oeed:
@lyqyd I get what you mean. I guess most of it is just differing philosophies, so I won't argue over that.

However, we really need examples of what sort of uses this has. If modifying the files stored here will cause crashes or problems, which I'd imagine might be quite likely, then I'd say that's certainly a good reason to hide it.

demhydraz:
```
matheus sleipnir ~ ⚡ ls /tmp                                     :( @ [8:43:13]
crontab.xBfbHl
systemd-private-03a627d4f97d43cd801b1b23fe5a1cb9-colord.service-lmgkGS
matheus sleipnir ~ ⚡ sudo rm -rf /tmp/*                          :( @ [8:43:18]
matheus sleipnir ~ ⚡ ls /tmp                                     :) @ [8:43:47]
```

ghost:
> However, we really need examples of what sort of uses this has. If modifying the files stored here will cause crashes or problems, which I'd imagine might be quite likely, then I'd say that's certainly a good reason to hide it.

Actually the point is that the files here are junk. They don't really need to be referred back to, and it should be treated as if the files could change or disappear at any time.

But maybe that's a bad idea, and maybe we should do something different to prevent things from changing when they need to be referred back to (although I would argue that it misses the point of the directory)

oeed:
That's a good point @cyanisaac. Ultimately hiding it is a matter of how much control you as the developer believe the user should have. Clearly, some people want it to be visible and others not.

As such I'd suggest we make `/tmp` standard but mention that it is permissible for the developer to treat it as a hidden folder if they wish (although the requirements for it to be hidden are that it is in the root folder, not just any folder called 'tmp').

ghost:
> That's a good point @cyanisaac. Ultimately hiding it is a matter of how much control you as the developer believe the user should have. Clearly, some people want it to be visible and others not. As such I'd suggest we make /tmp standard but mention that it is permissible for the developer to treat it as a hidden folder if they wish (although the requirements for it to be hidden are that it is in the root folder, not just any folder called 'tmp').

Good thinking, I will do something like that with my standard tomorrow (super late at the moment).

oeed:
Any update on this @cyanisaac? Once those changes have been made you can post it on the forums and soon after we can merge it :smile:

ghost:
I will get to work on this now. I've been busy this week that's why there's been a lack of updates :P

ghost:
More changes have been made. Let me know if I've missed anything.

oeed:
Looks good to me, anyone else have any issues?

demhydraz:
:+1: here

viluon:
@cyanisaac while you are saying that the temporary directory is located at `/tmp`, in the OS-side example you are using `/.tmp`. Either change that or list `/.tmp` in the "Location" field of your quick info table

Since you've removed the dot entries in https://github.com/cyanisaac/CraftOS-Standards/commit/c18adcdebb9f0dda209e18a8dbf49c8c6ddaba62, I think you just forgot about the OS example :wink: 

Other than that, I think this proposal is ready to be accepted!

Luca0208:
:+1: from me

ghost:
Gah I totally missed that. Fixed in OS-sided example.

viluon:
Thanks for the fix @cyanisaac :smiley:

lyqyd:
Looks good to me.  The standard appears to allow for /tmp to exist only in RAM, which is good!

oeed:
Okay, @cyanisaac this looks good to go. Can you post a topic in General with a link to the most up to date preview of the file (i.e. [this](https://github.com/cyanisaac/CraftOS-Standards/blob/56ed495ca4e428b9dad98ff1d743b83ca874d81a/Miscellaneous-Standards/Temporary%20Directory.md)) and another link to this page with a brief description of what the format is?

Once there are no objects or changes we'll then be able to merge!

ghost:
I edited my original thread, forgot to tell you guys. The links now link to these pull requests and some other stuff. Can it be merged?

Sorry for taking forever. I'm bad at remembering things.

oeed:
Oh of course, sorry @cyanisaac! You posted that ages ago!

Looks all good.

Issue #42, opened 2016-02-02
Standard & API writing order


oeed:
Although I had originally stated that the API should be written before the pull request is made I'm not so sure it's the right process. Instead I think that the API should be written _after_ the standard has approval. The PR can then have a `awaiting api` label assigned to it along with `accepted`.

viluon:
In that case, I'd suggest that the author should submit 2 PRs, one with the proposal (`proposal`) (which won't be merged, but you need to store the associated Markdown docs somewhere) and when that's accepted,(s)he submits another one (`standard`) with the actual standard and utilities.

viluon:
What do you think of that @oeed ?

oeed:
Hmm, actually that might work. Although, it separates things a bit too much I feel.

viluon:
Agreed, although cross-referencing the PRs might help with that..

lyqyd:
What's the need for a second PR?  May as well make any necessary changes on the initial one, and simply not pull until it's ready.  I'm having trouble seeing what the added benefit is to using a separate PR for the final standard text.

oeed:
Yeah I'd agree with @lyqyd, @viluon.

viluon:
Well, I know that 2 PRs is silly @oeed, but as it sort of "works" now, people open PRs when they've got nothing completed really... I'd like to fix that, just don't know how.

oeed:
I don't see a huge issue with the current PRs. Sure, it'd be nice if we could merge them faster, but placing too many barriers will just deter people from contributing.

PR #43, opened 2016-02-02
Metadata System

oeed:
This is a metadata based somewhat upon Lyqyd's previous proposal and previous input from others.

In simple terms, the system consists of a '.metadata' file stored in each folder which stores the metadata for each file in that folder as well as the folder's own metadata. The data is stored as Binary for fast and efficient storage.

I can see there being disagreements about this, so as with my suggestion in #42 I do not yet have an API written.

### DIFF
--- a/File Formats/File Metadata.md
+++ b/File Formats/File Metadata.md
@@ -51,6 +51,13 @@ To allow for quick, efficient and reliable MIME detection, each extension can on
 ### Unknown MIME
 
 If the MIME cannot be detected `nil` must be used as value. When loading the metadata of a file where the MIME is `nil` you must always try to redetect its MIME.
+
+## New file handling
+
+If a new file is created or a file without metadata or a MIME (i.e. non-nil) then default metadata must be generated.
+
+All of the required values (see **Keys**) *must* be set and optional values may be set if desired. All `Timestamp` values must use the default value of `1451606400` (00:00 1/1/2016 UTC). The MIME of the file must be detected based upon the extension, or, if a file is being created, set by the creating program.
+
 ## Metadata file contents
 
 The information in the *.metadata* file is the metadata for the folder itself and the files within the folder.
--- a/File Formats/File Metadata.md
+++ b/File Formats/File Metadata.md
@@ -1,0 +1,119 @@
+# File Metadata System
+
+## Folder structure
+
+Every folder (including root, discs and empty folders) has a file called *.metadata*. 
+
+For example, consider this file tree:
+
+```
+= documents
+	- image.ucg
+	- readme.txt
+	= code
+		- paint.lua
+= system
+	- startup
+
+```
+
+Under this metadata system, the file tree would be as such:
+
+```
+- .metadata
+= documents
+	- .metadata
+	- image.ucg
+	- readme.txt
+	= code
+		- .metadata
+		- paint.lua
+= system
+	- .metadata
+	- startup
+
+```
+
+It is important to note that any file with the name *.metadata* **must not be given metadata/have metadata stored**.
+
+### Behaviour and Treatment of *.metadata* Files
+
+Essentially, the user should never be able to change or see a *.metadata* file.
+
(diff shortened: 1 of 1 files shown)


oeed:
@demhydraz can you elaborate?

viluon:
@oeed I think that @demhydraz would like to have custom keys, which you disallow:

> It it **completely prohibited** to use any key not listed above or use an incorrect `Value`.

I would argue that for a solid metadata system, custom keys are a must. In @demhydraz's case, he would have to create _another_ metadata format to function alongside this one to store additional information his OS needs for proper handling of executables, for example. Information about what rights different user groups hold for the file is crucial to the shell, but this metadata standard is infeasible for its storage.

Feel free to get some inspiration ([[1]](https://en.wikipedia.org/wiki/Metadata_standard), [[2]](https://en.wikipedia.org/wiki/Metadata_modeling)) @oeed

demhydraz:
@viluon Thanks, man! That's exactly what I meant.

oeed:
Okay, I certainly understand where you are coming from. I was originally thinking that it would be a good idea but I can see a few issues.

The main issue is the inconsistency it will cause. Say, for example, @demhydraz starts using the key 'J' as a number value, say, the number of times it's been opened. If a folder then ends up @viluon's system which also uses the key 'J' as a number value, but instead for the colour depth, problems arise.

@viluon's system will try to open an image from the folder and use the number of times it has been opened, say, 7, as the colour depth. This has a reasonable chance of cause the file to become unreadable on that system.

It also requires a change to the format somewhat, as the keys are currently fixed you don't need a byte specifying what type it is as it already knows. That's not a major issue, but just something to consider.

demhydraz:
Then you add a 'XAttr-List' key.

oeed:
Once again, elaborate?

demhydraz:
Extended Attribute List. It's a path that operating systems can read to get additional keys.

viluon:
Wow, extended attr. list is a good idea @demhydraz! I was thinking that you might add something like a custom key key, custom key type key and a custom value key (but the format probably doesn't support multiple entries of the same type right?). That would allow you to set up "virtual keys", which would be formed by `custom-key = (string) "myKey"`, `custom-value-type = (string) "number"`, `custom-value = (number) 16`. That means 3 new keys would be added, with which you could express any value you'd like to @oeed.

**EDIT**:
Ehm, that might have confused you a little. Essentially you would add 3 new keys to the format, "custom key" (let's say K), "custom value type" (T is taken for MIME, so let's say it's lower case 't') and "custom value" (once again, V is taken, use 'v'). To add a _new key_, you would specify its name with K, its supposed type with t and its value with v. So for each new key you would add 3 new entries (that requires the keys in the file to be ordered, so new keys of the same name (K, t, v) can be reused for creating additional custom keys) which would describe the custom key.

Example:
I would like to add a custom key "awesomeness" which is a scale from 1 to 10 describing how awesome the file is. 
That means 3 new entries:
`K = "awesomeness"`
`t = "number"`
`v = 9`

Now I'd need another key, to describe how awesome the author of the file is. 3 more entries:
`K = "author's awesomeness"`
`t = "number"`
`v = 10`

The keys in the file have to maintain their order, because otherwise you wouldn't know which value belongs to which key (in this case, the types _can_ be swapped, because they're identical).

While usable, this is a rather crappy implementation of custom keys, mainly building atop what's already there. Extended attribute list as suggested by @demhydraz is a very good idea, but I would say the format itself should support custom keys in some way.

oeed:
I really don't get this extended attribute list, can you explain @viluon?

viluon:
@oeed extended attribute list would be a new key containing a file path where more metadata in a different format would be located. I believe [this](https://en.wikipedia.org/wiki/Extended_file_attributes) is the relevant Wikipedia page.

oeed:
After a discussion on Gitter, it has been proposed that the keys are changed to 3 characters with support for custom keys and extended attribute list support be added to allow for greater flexibility.

Issue #45, opened 2016-02-07
Communication Standard Template


oeed:
A template is required for communication standards. The [existing template](https://github.com/oeed/CraftOS-Standards/blob/master/Standard%20Proposal%20Guidelines.md) needs to be renamed to indicate that it is specific to file formats. There should also be a templates folder, unless there's some way you've integrated these in on `tutorial` @viluon?

ghost:
@oeed ooops I was so dumb, I  wrote one into the rednet pull request, it is in there...

ghost:
oh, this is an issue, not a pull request, I´m not very familiar with Github yet.

oeed:
No problem! Yeah, so for this project, issues are basically TODOs that (basically) anyone can do whereas pull requests are changes that need to be approved.

viluon:
Agreed @oeed, I'll rewrite templates in `tutorial`. I'll post updates about the progress here (will change from `todo` to `work in progress` when I actually start committing).

oeed:
To be honest @viluon I really think we need to get `tutorial` merged ASAP, it's leaving `master` lagging too far behind where we need it.

Can we maybe merge it and instead make smaller branches?

viluon:
I was just thinking about that, let's talk on [Gitter](https://gitter.im/oeed/CraftOS-Standards) @oeed (this is an unrelated Issue).

ghost:
I´ve got an idea for templates, see #46

oeed:
As mentioned in #47, there is quite a difference between _transmission methods_ and _protocols_. Hence, we actually need _two_ different templates.

ghost:
@oeed I think Rednet in a protocol too, but we should add another header Table Row, which contains Information if it can be used as Transmission Method. And Rednet is currently the only  ProtocolAndTrasmissionMethod thing. (once its accepted, I´ll release some Protocols based on Rednet)

oeed:
I might be wrong, but I feel like a protocol is more about _data_ being sent, not the method being sent. But I can see what you're saying.

@demhydraz or @viluon do you have any thoughts on this?

demhydraz:
> As mentioned in #47, there is quite a difference between transmission methods and protocols. Hence, we actually need two different templates.

This is right. Rednet is a transmission method. And something that'd send diffs along for file updating (say, like rsync/ftp) is a protocol.

oeed:
[...]

PR #44, opened 2016-02-07
Rednet Documentation

ghost:
Adds Rednet Standard Documentation as a base for other Communication Standards

PR #47, opened 2016-02-07
Rednet Protocol

ghost:
here I restert the pull request for the rednet prorocol

### DIFF
--- a/Communication-Protocols/rednet.md
+++ b/Communication-Protocols/rednet.md
@@ -1,11 +1,11 @@
 # Rednet #
 
-protocol | Rednet
+Protocol | Rednet
 -------- | ------
-transmission method | Modem (Peripheral)
-full name | Rednet
-short name | RN
-protocol ID | `rednet`
+Transmission Method | Modem (Peripheral)
+Full Name | Rednet
+Short Name | RN
+Protocol ID | `rednet`
 
 Rednet is a standard protocol included in computercraft. It is intented to be used as transmission method for other protocols.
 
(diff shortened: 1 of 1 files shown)


viluon:
`meta` This PR is a fix of the mess at #44.

oeed:
Just a few formatting fixes @justsomeoneelseagain:
- All the table column and row names need to be capitalised (i.e. 'Transmission Method' not 'transmission method') and the other cells should begin with a capital letter (i.e. 'The target computer...' not 'the target computer...')
- ~~Any number values, variable types and variable names need to have backwards ticks `around them so they're formatted like this:`65535`~~ Nice work, I see you've actually already done that since I wrote this, maybe just add it around things like `table` and `number` too.

Again, I'm not at all knowledgable about Rednet, but can you explain the 'short name' and 'protocol ID'? Are they specified/used by anything particularly?

ghost:
@oeed protocol ID is invented by me and supposed to be something like a MIME-Type for protocols. The short name is just a short variant of the name and the long is only words e.g. `HTTP` - `Hyper Text Transfer Protocol` , but the commonly used Name is `HTTP`, and the protocol is always referred in programming as `http`

Luca0208:
This looks nice :+1:

oeed:
@justsomeoneelseagain okay, we might need to decide whether they're needed. I can see benefits of having them, but we need to get that communication template #45 going.

ghost:
@oeed , ok, short and long names are not really needed, but we need something to identify protocols (and it has to be a string)

I´ll work on a Template as soon as this got accepted and then PR some more Protocols

oeed:
I'd say remove Short Name but keep Protocol ID.

This won't be accepted until a template is made and accepted so we don't need to update it. If you're going to make it, base it heavily off the file format template and only change what's needed.

lyqyd:
> "The data to be transmitted, specified by the protocol"

Should this not be the sender specifying the message?  I'm not even sure what you're trying to say here.

This is also not a great naming scheme for this standard, since one would want to create standards for individual protocols intended to be transmitted over rednet, like TRoR.  We can call both of them protocols, but we then ought to make distinctions between transmission/transport protocols (like Rednet, GPS, or LyqydNet) and data protocols (like TRoR).

oeed:
Yes actually that's quite true @lyqyd, this isn't a protocol but a method of transmission. @justsomeoneelseagain I'd just remove any references where you have called this a protocol. This also means removing the protocol ID.

ghost:
I agree, Rednet isn´t a Protocol, the whole document has to be written new from scratch, I won´t do that.

PR #50, opened 2016-02-09
Configuration File Standard

Luca0208:


bauen1:
From what i can see your using
`Key: Value`
But some people may also use
`Key = Value`

Luca0208:
Hmm, that's true, let's see what other people thing.

ghost:
ComputerCraft Introduced an Standard itself, using your Standard is like using something else instead of Rednet (but with modems too)

viluon:
@Luca0208 a couple of things:
- Please name your standard
- What are the benefits of using your standard over serialized Lua tables #41?
- Please use proper highlighting for code samples

Luca0208:
1.I'm not good with name, everyone feel free for suggestions
2. It is more readable than a normal Lua table
3. I will fix it

oeed:
To be honest I think config files should just use `textutils.serialise`. I just don't think people are going to bother using the format when there's very little difference and one has built in support.

Luca0208:
My point is, if someone who is new to CC sees these two files(can't use codenlocks currently, sorry):

```
Name:"Player1"
Difficulty:1
```

and

``` Lua
{
  Name="Player1",
  Difficulty=1,
}
```

The potential to screw up the bottom version is higher.

`meta`: Added code blocks for you ~ @viluon

viluon:
In that case, you are better off making a GUI settings screen. Nice people make those even for experienced users, so that they _don't have to_ dig through the config files.

I feel like closing this because #41 is superior. Any objections?

lyqyd:
Yes, this should be closed in favor of #41. Intentionally not supporting line breaks isn't fantastic, and valid-lua configuration files are easier to handle than plain text, within the context of ComputerCraft.

PR #51, opened 2016-02-09
Add Security Standards Category

Luca0208:
This might be good to prevent insecure programs.

### DIFF
--- a/Security-Standards/README.md
+++ b/Security-Standards/README.md
@@ -1,0 +1,7 @@
+# Security Standards
+
+These standards are security standards for in-cc
+
+## Proposal Requirements
+
+Contributors must declare your standard as secure.


lyqyd:
The only useful 'security standard' for ComputerCraft is: "Don't store sensitive data in ComputerCraft."

Anything else is better covered elsewhere. Just like one shouldn't roll their own crypto, making security recommendations is best left to the professionals.

Luca0208:
Well, if you have something like a password lock, using sha256 might be a reasonable standard.

viluon:
While this category is a good idea, there is no need to create one when there are no standards for it. Besides, your README is short and unusable (why should contributors think that your proposal is safe? Go ahead and make an insecure standard, it will fit into the category anyway (good luck getting it `recommended` though)). Rejected.

@Luca0208 also, "Using SHA256" is not a standard.

PR #52, opened 2016-03-05
Configuration Standard

lewisakura:


lyqyd:
If there is no standardization of the format for the contents of the file, I don't see any use to this. It's essentially just saying, "sometimes, a plain text file might be used for configuration data, but there's no way to determine whether or not that's the case for any given file". Since that's already the case, there's nothing added by including this as a standard.

viluon:
As @lyqyd said, pointless.

lewisakura:
What is the 'cock-up' label supposed to mean? The 'rejected' label is good enough...

lewisakura:
Anyway, redid the document. You might as well look at it.

viluon:
@LewisTehMinerz `rejected` is not enough in your case tbh. This is a [`cock-up`](https://duckduckgo.com/?q=cock-up&t=ffab&ia=definition) because 
- The 'standard' simplifies nothing
- You provide no utilities to deal with configuration files
- Some extensions you listed are in fact valid extensions to _existing_ configuration file standards, and you seem to completely ignore that
- There is no way of distinguishing a configuration file _as defined by your document_ from a plain text file (in fact, there is no way of distinguishing it from _any_ other file) -> **lacks specification**

> **Hint**: Standards kinda _are_ specifications

**EDIT**: Plus, as pointed out by @SquidDev on the Gitter chat, what @lyqyd pointed out is enough to state that your proposal is useless.

demhydraz:
I'd also like to add that your standard contradicts it's parent, _Plain Text_, by saying that config files should be in the .properties format (or as init files, or Lua tables - that's madness!).

_Plain Text_ says all text should be encoded as UTF-8, with Unix line terminators and no BOM, whereas the .properties spec says that all text should be encoded as Latin-1.

lewisakura:
You don't _need_ to follow what the .properties spec says. I've tried encoding it as UTF-8 and it still works the same.

lewisakura:
You know what? I'm just gonna drop this. I'm already sad as it is currently...

oeed:
Guys.... this was handled _reallllly_ poorly.

Rather than instantly shooting things down try and actually _support_ the contributor who has taken time out of their day to try and help. Don't go calling them stupid or instantly flagging them as cock-ups.

While, yes, the original proposal didn't really work, there is potential for a configuration file format.

Regardless, we want to _encourage_ contributors so this grows and becomes more useful by being nice and supportive. If you instantly make them feel like shit they're not going to want to contribute.

lyqyd:
Let's do away with the "cock-up" label entirely.  I can't see any reason to use it except to insult whoever created the work it's being applied to, and that will discourage valuable contributions along with less valuable ones.

I should have been more specific:  I don't see any use to this _in its current form_.  This would need to be modified to specify a format for the configuration data stored in the file, and would need some distinction between it and the `lua/table` format we're working on in #18.

oeed:
Agreed, `cock-up` is no more.

lewisakura:
Hmm, so you want it to be in _one_ format and not multiples?

viluon:
@oeed 

> Guys.... this was handled reallllly poorly.
> 
> Rather than instantly shooting things down try and actually support the contributor who has taken time out of their day to try and help. Don't go calling them stupid or instantly flagging them as cock-ups.

Agreed, sorry for that.

> While, yes, the original proposal didn't really work, there is potential for a configuration file format.

True and true, but I can see no configuration format proposal here.

> Regardless, we want to encourage contributors so this grows and becomes more useful by being nice and supportive. If you instantly make them feel like shit they're not going to want to contribute.

Again, apologies. And sorry for doing what I did @LewisTehMinerz, that was harsh.

However, this standard proposal is honestly useless. I mean

>  This would need to be modified to specify a format for the configuration data stored in the file

i.e. this would need to contain actual content. If you don't know how to make a standard @LewisTehMinerz, better don't do it (yet). Come over to the [Gitter chat](https://gitter.im/oeed/CraftOS-Standards) where we'll be happy to help you. Opening a PR with no idea and trying to come up with one on the go should be avoided at all costs imo, frankly I was a little bit offended by what you did (I just feel like this sort of contributions wastes collaborators' time). I agree with @oeed that a configuration file standard would be useful.

Proposed solution: let this stay closed @oeed, @LewisTehMinerz come over to [Gitter](https://gitter.im/oeed/CraftOS-Standards) and let's have a chat about this. You can open a new PR when we agree on the config file spec details.

lyqyd:
> Hmm, so you want it to be in one format and not multiples?

Yes. One standard should specify one thing, be it a file format, a communication protocol, etc. A standard should fulfill a use case. For example, the TRoR standard provides a method to transport screen content over a network, which is helpful for use with remote shell programs. A configuration file standard would also need to address a different use case than the existing `lua/table` format addresses.

PR #53, opened 2016-04-10
A United Standard for To-Do Lists

MineRobber9000:
This may extend off of the `lua/table` format you were working on, but I just spent a night working on this and now it's finished.

### DIFF
--- a/File Formats/Documents/To Do List.md
+++ b/File Formats/Documents/To Do List.md
@@ -1,7 +1,7 @@
-# _Basic To Do List_
+# _To Do List_
 | Information	|																						                                            |
 |-------------|---------------------------------------------------------------------------------------|
-| Version     | 0.2.0                                                                                 |
+| Version     | 0.3.0                                                                                 |
 | Type		  	| Document file format														                                      |
 | MIME		  	| `text/tdl`																			                                      |
 | Extensions	| `.tdl`	                                                                              |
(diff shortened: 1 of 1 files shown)


lyqyd:
Thanks for the submission!

I think that this would require further fleshing out before it should be accepted. For instance, I don't see any provision for subtasks. It also appears that the language used in the standard text would need some editing to match the style used in other standards. Another nice feature might be partial completion states, denoting that some task is some percentage complete, or a fractional representation of a total goal.

oeed:
I'd agree with Lyqyd, it's some nice work but does need a little more.

Maybe instead of booleans you could use a number between 0 and 1, 0 being not complete, 1 being complete and decimals being partially complete.

CrazyPyroEagle:
The MIME should be in the `text` group, not the `basic` group.

MineRobber9000:
I've decided to just make this a basic To Do List format. This comes because I can't find a way to make subtasks without it breaking. Are subtasks **THAT** important? Anyhow, I've renamed it to "Basic To-Do List".

lyqyd:
Why not something like this for subtasks:

``` Lua
{
  {"Show off how minimalistic this looks.",1,},
  {
    "Find out the Doctor's name.",
    {
      {"Track down the Doctor", 1,},
      {"Ask the Doctor what their name is", 0,},
    },
  },
}
```

MineRobber9000:
Ooh, that might work. A lot simpler then what I was originally going for when it came to subtasks.

MineRobber9000:
image

Implemented into TDLViewer, looking good?

MineRobber9000:
TDLClient Discontinuation Blurb (ignore if you're not interested, just wanted to say this)

> I've decided to get rid of TDLClient. It was a mess of spaghetti-code anyways, and implementing subtasks into it would delay the addition of subtasks to the standard. While the standard will move to 0.3.0, TDLClient will stay at 0.2.0 for the rest of the time it exists.

I'm going to update the TDLViewer pastebin to the 0.3.0 TDL standard (now with subtasks!) and update the standard document.

PR #54, opened 2016-05-08
Protocol Template

ghost:
Yes, I made it now, here is a Protocol Template, I've also made 2 Protocols ready (dns and chat), which i will create a pr, once this one is merged. (you can already see them in my other repository).

NOTE:
the "placeholder" is where the COS number is supposed to be.

### DIFF
--- a/Protocols/TEMPLATE.md
+++ b/Protocols/TEMPLATE.md
@@ -3,7 +3,7 @@
 placeholder | *Protocol Name*
 ------------|----------------
 Protocol Type | *P2P/Client-Server*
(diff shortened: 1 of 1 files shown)


lyqyd:
I can't really picture what you're trying to convey in your description of the packet structure. You might consider adding examples of how a packet type carrying a string and a packet type carrying a table would be specified.

It would also be helpful to clean up the numerous spelling and grammatical issues with the template  text.

Standards texts should never contain lists of implementations.

Thanks for the submission!

PR #58, opened 2016-10-21
Improve formatting of Standard Proposal Guidelines

SquidDev:
- Replace tabs with spaces
- Add COS number requirement
- Add "Quick information" header to remain consistent with other standards.
- Reformat several standards to be correctly aligned.

Several inconsistencies I've noticed:

[Universal Compressed Graphics](https://github.com/oeed/CraftOS-Standards/blob/master/File-Formats/image/ucg.md) takes a very different form to any other standard. I'm fine with this: it is very well written, but we may want to adjust the guidelines to account for this.

[TemporaryDirectory](https://github.com/oeed/CraftOS-Standards/blob/master/Miscellaneous-Standards/Temporary%20Directory.md) uses a different set of fields to any other standard: namely it is lacking a version field.

I also don't know if we want to define a specific column width that documents must be limited to.

### DIFF
--- a/File-Formats/TEMPLATE.md
+++ b/File-Formats/TEMPLATE.md
@@ -1,18 +1,21 @@
-**This file is a template of how a standard proposal document should look like.**
+**This file is a template of how a file format standard should look.**
 
 ---
 
-# _Standard Name_
-| Information	|																						                                            |
-|-------------|---------------------------------------------------------------------------------------|
-| Version 		| [semver](http://semver.org) version tag												                        |	
-| Type		  	| Category																				                                      |
-| MIME		  	| `mime/type`																			                                      |
-| Extensions	| `.extension1`; `.extension2` (Use `none` if your standard does not use file extensions)	|
-| Deprecates	| (List) Names of standards deprecated by _Standard Name_				                  			|
+# *COS num:* _Standard Name_
+
+## Quick information
+| Information |                                                                                        |
+| ----------- | -------------------------------------------------------------------------------------- |
+| Version     | [semver](http://semver.org) version tag                                                |
+| Type        | Category                                                                               |
+| MIME        | `mime/type`                                                                            |
+| Extensions  | `.extension1`; `.extension2` (Use `none` if your standard does not use file extensions)|
+| Deprecates  | (List) Names of standards deprecated by _Standard Name_                                |
 
 ### Technical Details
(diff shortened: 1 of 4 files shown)


lyqyd:
I'm marking this as accepted, since it corrects one oversight and makes no other functional changes. I don't think it'll require much conversation to be merged, but I'd like to give @oeed and @viluon a chance to disagree or toss a thumbs-up in here. If we haven't heard from them in a couple days, I'll go ahead and merge it.

I'd appreciate it if you could open an issue for the two inconsistencies you've noticed so that they can be talked out as necessary. The temporary directory could also have a pull request submitted adding 1.0.0 as the version number, if you're feeling like it.

Thanks for the contribution!

oeed:
No complaints here. I feel like converting UCG to the correct format would be useful though. It was written before this repo existed by someone else, so naturally didn't use the format.

viluon:
Sorry for the delay guys, I agree with @oeed. Thanks for the heads-up @SquidDev!

SquidDev:
I've just noticed that the [file format template](https://github.com/oeed/CraftOS-Standards/blob/master/File-Formats/TEMPLATE.md) is also slightly different: would you like me to fix that up in this PR too?

lyqyd:
Yeah, that seems reasonable to fix on this PR as well.

SquidDev:
Having a look at the file format template: would it be acceptable to delete it? It is identical to the standard proposal guidelines and so I see little reason to have both.

lyqyd:
You're right that they're identical.  I think a better course of action would be to re-visit the Standard Proposal Guidelines document and figure out what we want to do with it to generalize it, or determine if it needs to be kept. It's probably useful to have templates for the various categories of standards, but the overarching guidelines shouldn't be specific to one type of standard.

I think that one's outside the scope of this PR, though.

lyqyd:
> I've just noticed that the file format template is also slightly different: would you like me to fix that up in this PR too?

@SquidDev Was this something you still wanted to add to this PR, or would you like to have this one merged and submit a separate PR? I'm fine with either route.

SquidDev:
@lyqyd Sorry, I've been a bit behind on everything. I've updated the template now. Is everyone happy with this?

lyqyd:
Thanks, @SquidDev! Didn't know if that change had slipped past you.

This PR looks great to me. I'm happy to merge it, unless there are any objections?

PR #63, opened 2017-01-10
DNS Protocol

ghost:
This is adding a DNS Protocol Description according to the Rednet API.

### DIFF
--- a/Protocols/dns.md
+++ b/Protocols/dns.md
@@ -1,0 +1,41 @@
+# DNS
+
+## Quick Information
+
+| Information |                                                                |
+| ----------- | -------------------------------------------------------------- |
+| Version     | 1.0.0                                                          |
+| Type        | Protocol                                                       |
+
+## Technical Details
+
+DNS defines a method of looking up hostnames in the context of protocols.
+
+Any computer can broadcast a `lookup` for a hostname/protocol pair and will receive a `lookup response` from computers who identify themselves with the corresponding hostname/protocol pair.
+
+Because it's not just hostnames but hostname/protocol pairs, there can be several computers using the same hostname for serving different protocols.
+
+A computer can identify himself with every hostname/protocol pair he wants and with as many hostname/protocol pairs as he wants, as long as there are no duplicated hostname/protocol pairs in the world. The computer HAS TO lookup a hostname/protocol pair before identifying with it, if it is taken, the computer is NOT ALLOWED to identify with it.
+
+Packets are composed of 3 strings:
+* the packet type
+* the hostname
+* the protocol
+
+These strings are combined in a table:
+
+```
+  {
+  sType = "<packet type>"
+  sHostname = "<hostname>"
+  sProtocol = "<protocol>"
+  }
+```
+
+### Looking up a hostname/protocol pair
+
+* The uplooking computer broadcasts a packet of type `lookup`, where the hostname and protocol string are the hostname/protocol pair to look up.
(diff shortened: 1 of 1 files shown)


ecktec:
What do I have to do in order to get this merged?

oeed:
Does this follow Rednet exactly? I think somethings do need to be clarified a bit such as what channel it is on and how to use it rather than a fairly broad outline.

ecktec:
I read it out of the rednet API source code.
There isn't any channel mentoined, because it uses rednet.broadcast().
I really don't know why I do this, because nobody would ever want to write somethig that does that and rather use the rednet API, so this should just be a specification.

ecktec:
The source code from the rednet API:

lookup function:
```lua
function lookup( sProtocol, sHostname )
    if type( sProtocol ) ~= "string" then
        error( "expected string", 2 )
    end

    -- Build list of host IDs
    local tResults = nil
    if sHostname == nil then
        tResults = {}
    end

    -- Check localhost first
    if tHostnames[ sProtocol ] then
        if sHostname == nil then
            table.insert( tResults, os.getComputerID() )
        elseif sHostname == "localhost" or sHostname == tHostnames[ sProtocol ] then
            return os.getComputerID()
        end
    end

    if not isOpen() then
        if tResults then
            return table.unpack( tResults )
        end
        return nil
    end

    -- Broadcast a lookup packet
    broadcast( {
        sType = "lookup",
        sProtocol = sProtocol,
        sHostname = sHostname,
    }, "dns" )

    -- Start a timer
    local timer = os.startTimer( 2 )

    -- Wait for events
    while true do
        local event, p1, p2, p3 = os.pullEvent()
        if event == "rednet_message" then
            -- Got a rednet message, check if it's the response to our request
            local nSenderID, tMessage, sMessageProtocol = p1, p2, p3
            if sMessageProtocol == "dns" and type(tMessage) == "table" and tMessage.sType == "lookup response" then
                if tMessage.sProtocol == sProtocol then
                    if sHostname == nil then
                        table.insert( tResults, nSenderID )
                    elseif tMessage.sHostname == sHostname then
                        return nSenderID
                    end
                end
            end
        else
            -- Got a timer event, check it's the end of our timeout
            if p1 == timer then
                break
            end
        end
    end
    if tResults then
        return table.unpack( tResults )
    end
    return nil
end
```

hosting mechanism:
```lua
local tHostnames = {}

function host( sProtocol, sHostname )
    if type( sProtocol ) ~= "string" or type( sHostname ) ~= "string" then
        error( "expected string, string", 2 )
    end
    if sHostname == "localhost" then
        error( "Reserved hostname", 2 )
    end
    if tHostnames[ sProtocol ] ~= sHostname then
        if lookup( sProtocol, sHostname ) ~= nil then
            error( "Hostname in use", 2 )
        end
        tHostnames[ sProtocol ] = sHostname
    end
end

function unhost( sProtocol )
    if type( sProtocol ) ~= "string" then
        error( "expected string", 2 )
    end
    tHostnames[ sProtocol ] = nil
end
```

part of run function dedicated to dns:
```lua
elseif sEvent == "rednet_message" then
		    -- Got a rednet message (queued from above), respond to dns lookup
		    local nSenderID, tMessage, sProtocol = p1, p2, p3
		    if sProtocol == "dns" and type(tMessage) == "table" and tMessage.sType == "lookup" then
		        local sHostname = tHostnames[ tMessage.sProtocol ]
		        if sHostname ~= nil and (tMessage.sHostname == nil or tMessage.sHostname == sHostname) then
		            rednet.send( nSenderID, {
		                sType = "lookup response",
		                sHostname = sHostname,
		                sProtocol = tMessage.sProtocol,
		            }, "dns" )
		        end
		    end
```

PR #64, opened 2017-01-20
Minor cleanup in a couple of places

### LINKED ISSUE #60: Formats missing the version field.
The following formats are missing a version field:
- [Temporary Directory](https://github.com/oeed/CraftOS-Standards/blob/master/Miscellaneous-Standards/Temporary%20Directory.md)
- [Directory](https://github.com/oeed/CraftOS-Standards/blob/master/File-Formats/application/directory.md) (this is in a different directory to the above - is this correct).
- NFT and Paint
- Lua and Plain text

I'd presume we'd want to use 1.0.0 for all of these? I'm happy to put together a PR once #58 is merged.

### LINKED ISSUE #61: File naming – should we include COS numbers?
Just wanting some thoughts on whether we should include COS numbers in the file name of standards. For example, `lua.md` would become something along the lines of `COS 2 - lua.md`.

Also, regarding file names that are acronyms (i.e. `tror.md`) should be also include the expanded name (i.e. `tror - terminal redirection over rednet.md`).

I can see the arguments for and against both, so interested to know what everyone else thinks.

SquidDev:
## Currently added
 - Add version field to all remaining standards
 - Move temporary directory to be in the same folder as the normal one.
 - Add MIME type to temporary directory
 - Remove file specific fields from `Tutorial/StandardProposalGuidelines.md`
 - Could we move `File-Formats/README.md` to the main ReadMe and have a list of all standards (with COS numbers).
 - Some files are `LikeThis.md` or `Like-This.md` or `like this.md`. What would we prefer?
 - `File-Formats/README.md` uses different names from the actual standards.

## To consider
 - ~~The temporary directory could have a better mime type.~~
 - " Images that only require 1 colour per pixel" doesn't make sense (in `File-Formats/README.md`).
 - `temporary directory.md` has a different type from `directory.md`.

Closes #60 and closes #61.


lyqyd:
I'm honestly a fan of `like-this.md`, since I don't much care for either spaces or capital letters in file names where they aren't necessary.  I don't have a strong preference on this point, anything is fine with me as long as it's consistent throughout the repo.

I don't have any good suggestions for either the temporary directory MIME type or the file format readme related issues.

SquidDev:
I quite like `like-this.md`. I'm happy to rename everything (and possibly add #61) at the same time, if people agree?

dmarcuse:
I like that as well.

lyqyd:
So, something like `cos-42_like-this.md`, perhaps? Hyphens between words for each part of the file name, underscore to separate the COS number from the rest.

SquidDev:
I'd prefer to leave off the `cos-`. I'm not a great fan of the underscore but I guess it makes sense. Should digits be two numbers (`06-ntf.md`) or just the minimum number required? I'd prefer the former as then proposals appear in order, but it isn't really extensible if we ever get more than 100 standards.

oeed:
I'm all for continuity with file names. I don't quite understand why you'd remove spaces and capital letters though, these are documents not code. Readability is more important than being able to use the name from command-line. 

I do fully support having the COS number in the file name though, it makes it easier to find the one you want.

viluon:
`alexander-hamilton.md` has my vote. As to the COS numbers, placing them in front of the file name would likely order any directory listings by them - is that something we want? I don't really see how is the COS number useful...

If added, they definitely shouldn't be in a non-forward-compatible format (such as `06-ntf.md`).

The more I'm thinking about this, the less I'm sure a directory structure like the one we went with is feasible. I'm starting to think a database would suit much better, or something like a directory with all the standards named `cos-*number*.md` and a separate folder with categories, where each category is a list of what is "tagged" as belonging there, with links to the standards in question...

viluon:
Regarding the [Temporary Directory](https://github.com/oeed/CraftOS-Standards/blob/a8a60d58f68e931e83c5682e0899fcac34be52ed/Miscellaneous-Standards/Temporary%20Directory.md) standard, it should have the MIME type of a [directory](https://github.com/oeed/CraftOS-Standards/blob/5f0f699e11ad870cad89a0d1248c3433816ac4ae/File-Formats/application/directory.md). What is there so special about it, anyway? It is stored and operated with in the same manner. I'm strongly against different MIME types for directories (and any useless metadata in general).

SquidDev:
> I'm starting to think a database would suit much better, or something like a directory with all the standards named `cos-*number*.md_` and a separate folder with categories, where each category is a list of what is "tagged" as belonging there, with links to the standards in question...

I don't disagree with this, though I'd rather also include the standard name too: it would make it easier when searching for things (such as on a local copy, when you won't go through the readme).

viluon:
`meta`: I don't know where to put this (sorry for another comment), feel free to yell at me should you think I ought to open an issue instead.

`on-topic`: These "Submission Checklist" items shouldn't technically be a required condition for a standard to be accepted:
> If your standard gets accepted, will its users benefit from it? Does your proposal help solve frequent issues related to the category it should be used in? Does it help with compatibility (esp. backward and forward compatibility with future standard revisions)? Is it easy to use?

> Is your proposed standard easy to use? Does it require third-party code or brand new code to be written in order to be used properly, or do you pack needed utilities with the proposal? Are these utilities efficient? 

What I mean is that our main goal should be to make the standard definitions unambiguous and (technically) provide hosting for them. Just because the base idea of a standard is crappy or we disagree with it doesn't mean it shouldn't be accepted.

SquidDev:
What are people's opinions on @viluon's comments? I want to know people's thoughts before I commit (to) anything.

lyqyd:
I like the idea of the restructuring of the folder and going with all standards in one directory, named by COS number.

I also agree that we can completely scrap the easy-to-use requirement for standards. I do think that it's important to ensure that the scope of proposed standards is broad enough that they can be useful to multiple developers.

oeed:
I agree with Lyqyd, although I'm of the belief that when it comes to file names we should use something like `042 - Format Name.md`

demhydraz:
I'm with oeed on the namey bits.

dmarcuse:
I prefer hyphens or underscores to spaces, no preference on number format.

`format_name.md`

viluon:
I really wouldn't prefix the COS number with zeroes - what good will it do? Regarding the name, I think underscores or hyphens win over spaces, they're easier to deal with from a command line and (should anyone want to do that) from the CraftOS shell.

oeed:
I take a punt at saying that the vast majority of people would not be using this from the command line, they'll probably just be reading them from this repo. Those who do will know how to escape spaces anyway. It just the file names more readable.

Prefixing the COS numbers with zeros allows them to be sorted by the number. Many systems don't 'properly' sort file names with numbers. i.e. it'd sort like this:

```
1 - File
11 - File
12 - File
13 - File
2 - File
24 - File
```

Whereas with zeros, it does this:

```
01 - File
02 - File
11 - File
12 - File
13 - File
24 - File
```

viluon:
I tested this in a private repo - GitHub is indeed an example of this naïve approach to directory listings.

The point of the approach I suggested, however, was that people wouldn't *need* to browse the files themselves, they'd open a category/tag Markdown file instead that would contain links to the actual standards. Essentially, the naming convention doesn't matter (as it's only for the links) as long as it is consistent and easy to use for the collaborators. Therefore, I'm for the no-prefix underscore/dash approach (`2-fancy-standard`).

Prefixing with zeroes places a top limit on the number of standards we can accept. A change of that limit has a worst case scenario complexity of *O(n)*, where *n* is the total number of standards already accepted (to add a zero to the prefix of each one of them), while no prefix is obviously *O(1)*.

CrazedProgrammer:
`2-fancy-standard` seems fine

SquidDev:
Right. [I went for it](https://github.com/SquidDev-CC/CraftOS-Standards/tree/hotfix/versions). Happy to revert the commit if people aren't happy with it. A couple of things I want to also ask:

### Wrapping text. 
Some files are horrible to read in an editor (unless I enable word wrap which I'm loath to do). It might be worth preferring a wrap limit (such as 120 characters) though sometimes we'd need to ignore that (like Markdown tables).

However, small changes to make the diff more dirty as the entire paragraph needs to be re-wrapped.

### Documentation
I've kinda bundled all the [tutorials and templates together](https://github.com/SquidDev-CC/CraftOS-Standards/tree/hotfix/versions/tutorial). Its a mess. I'm not sure I want to clean it up in this PR but it would be worth considering.

### And another thing
I've noticed that the standard is directory, but is referred to as folders within the spec. Do we need to clarify this?

dmarcuse:
But then this happens 😕 

Screenshot

SquidDev:
@apemanzilla Yes, and I realise that from a browsing point of view it is a problem. I'm hoping that most people will browse online and so will be able to use [the standards list](https://github.com/SquidDev-CC/CraftOS-Standards/blob/hotfix/versions/standards/README.md) instead, which *is* sorted numerically.

viluon:
@apemanzilla: @oeed and I [commented](https://github.com/oeed/CraftOS-Standards/pull/64#issuecomment-275551229) on the sorting issue.

@SquidDev Text wrap is probably a good idea.
I *think* meta docs will need a revamp anyway, I can look into them when I have time (2-3 months from now). Feel free to [open an issue](https://github.com/oeed/CraftOS-Standards/issues/new) to track this.
Regarding directory vs. folder, would you be so kind and point out where exactly the clashes are? I'd probably go with "directory" (there's a 30% chance I'm misunderstanding you, sorry if that's the case).

SquidDev:
> Regarding directory vs. folder, would you be so kind and point out where exactly the clashes are?

From the [`3-directory.md`](https://github.com/SquidDev-CC/CraftOS-Standards/blob/hotfix/versions/standards/3-directory.md) file:

> It is important to keep in mind that some systems, such as Silica, sometimes treat **folders** as single files.

Also worth noting that the type field is given by `Folder/directory` (compared with `Path / Directory` in the temp directory spec). The directory spec probably needs fleshing out but that is not the point of this spec.

viluon:
@SquidDev [this](https://github.com/oeed/CraftOS-Standards/blob/409add04c51a77cb6bbec367da058774919682d1/File-Formats/application/directory.md) (permalink) is the current version of the standard at the time of writing. From what I can tell, it's a rather badly written one, I would suggest making a new directory standard from scratch. That would keep this one around for historical reasons...

... Although I must say there doesn't seem to be a benefit to adopting a directory "standard" - it really doesn't define anything beside the MIME type. Handling of inodes is a matter of implementation in the file system. I'd standardize a file system rather than a general idea of a directory (not to mention this one is rather vague - wait, it doesn't even *define* what a directory is?!)

**TL;DR**: Ignore this. Should have been resolved at PR approval time.

dmarcuse:
@viluon I'm aware, it's just a minor nuisance.

SquidDev:
Do people have any feedback on the current state of the PR?

CrazedProgrammer:
Looks good

standards/README.md
```diff
@@ -0,0 +1,13 @@
...
+| 1   | [Plain Text](1-plain.md)                        | File format      | Simple text documents                                        |
+| 2   | [Lua](2-lua.md)                                 | File format      | Lua source code                                              |
+| 3   | [Directory](3-directory.md)                     | File format      | A collection of multiple files and directories               |
+| 4   | [Paintutils Image](4-paint.md)                  | File format      | Storing simple bitmaps                                       |
+| 5   | [Universal Compressed Graphics](5-ucg.md)       | File format      | Optimised format for storing images                          |
+| 6   | [Nitrogen Fingers Text](6-nft.md)               | File format      | Images with text support                                     |
+| 7   | [Temporary directory](7-temp-directory.md)      | File format      | Storing temporary files                                      |
+| 10  | [TRoR](10-tror.md)                              | Protocol         | Redirecting terminal state and interactions over Redent.     |
```
lyqyd: Typo to correct here; Red**en**t to Red**ne**t.

tutorial/file-formats.md
```diff
@@ -0,0 +1,40 @@
...
+If the file format is new it must only be given extension; there's no point have 3 different extensions for the same
+file.
+
+If the file format is an existing format *and it already* uses multiple extensions then you should list all of the
+commonly used ones.
+
+### MIME
+Example: `text/lua`, `image/sketch`, `silica/theme`
+
+Every file format must be given ***one*** MIME. If more than one already exists pick **one**.
```
lyqyd: Perhaps "exactly one" instead of ***one***? Precise language can't get lost through format stripping from copy-pasting the way formatting emphasis can.

tutorial/getting-started.md
```diff
@@ -0,0 +1,117 @@
...
+at [our Gitter chat](https://Gitter.im/oeed/CraftOS-Standards).
+
+#### A Note on Licensing
+All standards published here on CraftOS-Standards are licensed under
+the [GNU Free Documentation License](../LICENSE.md). **This includes all code examples**. For the sake of open source
+software and simplicity alike, your libraries should also be licensed
+under [similar FOSS licenses](http://choosealicense.com/). After all, licensing official utilities for an open source
+standard under a closed source license wouldn't make much sense.
+
+### Step 6: Almost There
```
lyqyd: Then where is the finish?

Perhaps this could be removed?

[Edit: This didn't end up including sufficient context; my original comment refers to line 116 and 117, which Github omitted.]

tutorial/getting-started.md
```diff
@@ -0,0 +1,117 @@
...
+stick to the code examples instead of writing an [API](https://en.wikipedia.org/wiki/Application_programming_interface)
+later on. If that is the case, all code required to deal with your standard has to be included in the
+documentation. Otherwise, write up code examples only for dealing with *the library* you will create in step 5. Keep
+them to the point, one per every scenario your standard should be used in. More thorough code should be kept elsewhere
+(you will be told where in just a moment!).
+
+One thing to note about example code is that it should be easily readable. Every important line should be commented
+(i.e. explain what you're doing!) and the code should be well formatted. Preferably, use camel case for all
+variables. If your code outputs something to the screen, add a comment with the exact same text. In Markdown, use proper
+highlighting for the source code (i.e. \`\`\`Lua ... \`\`\`).
```
lyqyd: Looks like this may need a lowercase L for the syntax highlighting to work?

CONTRIBUTING.md
```diff
@@ -1,25 +1,49 @@
...
+Welcome, fellow developer! First off, make yourself familiar with the [tutorial](./Tutorial/GettingStarted.md). It's a
+long read, so enjoy it with some :coffee:. If you'd like to create new standards, the tutorial should actually give you
+all that you'll need to know. Thanks for helping us out by standardizing something awesome!
 
 ### A Note on Licensing
-All standards published here on CraftOS-Standards are licensed under the [GNU Free Documentation License](../LICENSE.md). **This includes all code examples**. For the sake of open source software and simplicity alike, libraries for dealing with your standard should also be licensed under [similar FOSS licenses](http://choosealicense.com/). After all, licensing official utilities for an open source standard under a closed source license wouldn't make much sense.
+All standards published here on CraftOS-Standards are licensed under
```
lyqyd: Perhaps "licensed" should point to LICENSE.md, and the actual name of the license should point toward its [page on gnu.org](https://www.gnu.org/licenses/fdl-1.3.en.html), for those seeking more information on the license itself.

Terminology.md
```diff
@@ -1,94 +1,109 @@
...
 
-An executable is a file containing Lua (or otherwise) that can be loaded by CraftOS (or otherwise) and executed. Most executables are user-facing applications.
+An executable is a file containing Lua (or otherwise) that can be loaded by CraftOS (or otherwise) and executed. Most
+executables are user-facing applications.
 
 #### 3.2. Library
 
-A library is a file containing Lua (or otherwise) that can be loaded by CraftOS (or otherwise) and used for code reusal. Do not the difference between an API and a library.
+A library is a file containing Lua (or otherwise) that can be loaded by CraftOS (or otherwise) and used for code
+reusal. Do not the difference between an API and a library.
```
lyqyd: Should this be "note" in the last sentence?

lyqyd (commented):
Some of these questions may be outside of the intended scope of the PR; I'm fine with splitting one or more off into their own issues for discussion if necessary.

Looks good otherwise!

tutorial/file-formats.md
SquidDev: I'm not really happy with "If more than one already exists pick **one**." A possible alternative.

> If more than one MIME type is applicable then choose the one that best suits it.

SquidDev:
Thanks @lyqyd for the feedback and typo catches. I think any major rework of the guidance would be better for a separate PR but I'm happy to fix any minor issues with things.

tutorial/file-formats.md
lyqyd: Yeah, that seems like a good alternative.

tutorial/file-formats.md
```diff
@@ -16,7 +16,7 @@ commonly used ones.
 ### MIME
 Example: `text/lua`, `image/sketch`, `silica/theme`
 
-Every file format must be given ***one*** MIME. If more than one already exists pick **one**.
+Every file format must be given excatly one MIME. If more than one already exists pick **one**.
```
lyqyd: I hate to nitpick again, but I don't think that's *exactly* how you spell that word. ;-)

SquidDev: Serves me right for doing this late at night without spell check.

CONTRIBUTING.md
```diff
@@ -1,25 +1,49 @@
+# Guidelines for Contributing to CraftOS-Standards
 
-## Guidelines for Contributing to CraftOS-Standards
-
-Welcome, fellow developer! First off, make yourself familiar with the [tutorial](./Tutorial/GettingStarted.md). It's a long read, so enjoy it with some :coffee:. If you'd like to create new standards, the tutorial should actually give you all that you'll need to know. Thanks for helping us out by standardizing something awesome!
+Welcome, fellow developer! First off, make yourself familiar with the [tutorial](./Tutorial/GettingStarted.md). It's a
+long read, so enjoy it with some :coffee:. If you'd like to create new standards, the tutorial should actually give you
+all that you'll need to know. Thanks for helping us out by standardizing something awesome!
```
viluon: The "that" here is redundant.

CONTRIBUTING.md
```diff
@@ -1,25 +1,49 @@
...
+the [GNU Free Documentation License](https://www.gnu.org/licenses/fdl-1.3.en.html). **This includes all code
+examples**. For the sake of open source software and simplicity alike, libraries for dealing with your standard should
+also be licensed under [similar FOSS licenses](http://choosealicense.com/). After all, licensing official utilities for
+an open source standard under a closed source license wouldn't make much sense.
 
 ## Getting Started: Becoming a CraftOS-Standards Contributor
 
-If, for some reason, you'd like to join our amazing [contributor team](https://github.com/oeed/CraftOS-Standards/graphs/contributors), you will need to know a few more things about how this repository works. Great news! We wrote them down here so that you won't have to bother us about the details.
+If, for some reason, you'd like to join our
+amazing [contributor team](https://github.com/oeed/CraftOS-Standards/graphs/contributors), you will need to know a few
```
viluon: "Contributors" actually means everyone who has committed to the repo. Collaboratorship is the word GitHub uses for people with write access.

viluon: This appears a few more times till the EOF.

README.md
```diff
@@ -4,46 +4,82 @@
...
-7. **Discussion** – if anyone has any issues or improvements with your standard discuss them on the pull request and change the RFC's markdown file until people have agreed upon the standard. At this point the pull request will be merged, you standard has been accepted!
+1. **Make sure there isn't an existing standard** – the whole point of this repo is to ensure we don't have
+   two [slightly different standards](https://xkcd.com/927/); there should only be one. You can either check
+   the [PR page](https://github.com/oeed/CraftOS-Standards/pulls?q=is%3Apr) or directly browse
+   the [source tree](https://github.com/oeed/CraftOS-Standards/tree/master/) of this repository. You should also look at
+   the Submission Checklist (further down this page).
```
viluon: This point is rather questionable - I know that multiple standards dealing with the same issue are painful, but who's to decide which ones are too similar and which ones are not? It's not really an issue to keep many standards around, I mean, GitHub literally does that job for us. Not much growth in the number of CC standards is to be expected, anyway. I think we should focus on ensuring the proposals' unambiguity rather than generally good design.

README.md
```diff
@@ -4,46 +4,82 @@
...
+2. **Develop your standard** – think about the different use cases and try to create clearly defined standards and rules
+   for your standard. If applicable, it's probably a good idea to create an API or example code for your standard.
+3. **Write an RFC** – once you have created your standard create a Markdown document which clearly explains what your
```
viluon: Perhaps mention here that RFC stands for "Request For Comments", and stress that 3rd party feedback is crucial.

README.md
```diff
@@ -4,46 +4,82 @@
...
+   standard is, how it works and how to use it. If you're at all unsure of what to do, look at the RFCs of accepted
+   standards. Make sure that the document covers your standard extensively, and contains thorough examples. You want
+   your standard to be the most popular one and to be used by as many people as possible, so this step should take the
+   longest amount of time (not counting #7 Discussion).
+4. (Optional) **Create a GitHub repository for in-depth docs and example source code** – Markdown is great, but
```
viluon: I think we should accept GitLab and perhaps Bitbucket repos as well.

viluon: (That is, for [GitLab CE](https://gitlab.com))

README.md
```diff
@@ -4,46 +4,82 @@
...
+   documentation (such as a [wiki](https://help.github.com/articles/about-github-wikis/)) and/or example code that is
+   either too long to be included in the Markdown document or split into too many files and follows a specific directory
+   structure. Add the link to your newly created repository to the Markdown document from step 3.
+5. **Create a pull request** – create a pull request with your completed Markdown file. If you think it belongs in a
+   folder put it in there/create it, but we can easily move it if you're unsure. Any code examples should be embedded in
+   the Markdown file. Your pull request's description should be a explanation of the format and why it's needed, but not
+   too descriptive, that's what the Markdown file is for. Remember that the pull request isn't permanent, it is there to
+   agree upon the standard, the Markdown file is permanent and should have all the detail. However, any full APIs should
+   be uploaded to Pastebin and a link should be added to the Markdown file. *There should be no code files in the repo,
+   markdown only.*
```
viluon, line 47: Pastebin? Uh oh. Also, let's stress that we mean oeed/CraftOS-Standards, this wording could imply that "no code files" should be present in the repo from step 4.

README.md
```diff
@@ -4,46 +4,82 @@
...
+   your standard is (i.e. *an method redirecting a terminal over Rednet*). The post's title should also make it pretty
+   obvious it's an RFC (i.e. *RFC: TRoR (Terminal Redirection over Rednet)*). To keep everything centralised it's best
+   to keep discussion of the RFC on the pull request, *NOT* the forums post.
+7. **Discussion** – if anyone has any issues or improvements with your standard discuss them on the pull request and
+   change the RFC's markdown file until people have agreed upon the standard. At this point the pull request will be
+   merged, you standard has been accepted!
 
 If you think a pull request should already have been merged @ mention one of the moderators (see below).
 
 ## Submission Checklist
```
viluon, line 59: This whole section is questionable, as mentioned earlier.

viluon (changes requested):

viluon:
I'm not against your changes @SquidDev, the issues I found mostly come from the text that defined our conventions in the first place. I'd welcome @oeed's and @lyqyd's (and anyone else's!) feedback on this.

CONTRIBUTING.md
SquidDev: I'll fix this but honestly this entire thing needs a rewrite and I don't think this is the appropriate PR for this.

CONTRIBUTING.md
SquidDev: I'd also prefer to split this into a separate section: most people want to contribute, not become a collaborator. Anyway, separate PR.

README.md
SquidDev: > stress that 3rd party feedback is crucial.

Shouldn't this happen before you're written up the RFC? - Whilst developing your standard.

SquidDev:
I'd rather avoid any more phrasing changes to the existing text. I'm more interested in resolving the questions in the original PR comment than fixing what we have: that can be left til a later commit.

lyqyd:
Yes, I agree that we should separately re-evaluate some of the text in README.md, but that is outside the scope of a "minor cleanup", and could be worthy of a more in-depth discussion. I'm okay with accepting this PR as it stands.

CONTRIBUTING.md
viluon: Yeah, agreed :/

viluon (approved):

viluon:
You're absolutely right there @SquidDev @lyqyd, I just pointed out what I thought should be changed, ignoring the scope of this PR. Sorry.

**EDIT**:
As for the Miscellaneous category, I'm in favour of deleting.

SquidDev:
@oeed - do you have anything to add/change?

oeed (approved):
Looks fantastic, nice job!

oeed:
Is this ready for merging?

PR #66, opened 2017-06-22
Update TROR standard to include terminal palettes

SquidDev:
This defines three new codes:

 - `TM`: Sets a single palette's colour.
 - `TN`: Requests a palette's colour from the client.
 - `TO`: Carries a palette's colour to the server in response to `TM`.

Also adds an optional `palette` field to the `TT` packet.

### DIFF
--- a/standards/10-tror.md
+++ b/standards/10-tror.md
@@ -139,8 +139,9 @@ The table MUST contain the following fields:
 
 The table SHOULD contain the following fields. If they are ommitted then the
 client SHOULD revert to the default values.
-| ------------- | ------------------------------------------------------------ |
+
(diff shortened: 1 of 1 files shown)


dmarcuse:
Is it necessary to have both `TM` and `TO`, since they're identical in terms of structure?

SquidDev:
@apemanzilla Technically no, but the standard generally uses packets for exactly one thing - the cursor 
 position set/request/response packets work similarly.

SquidDev:
I realise the `TT` extension does not include support for palettes. Any suggestions on how this should be implemented? The simplest would be for the client to reset to the default palettes should they be missing.

PR #67, opened 2017-07-25
Make Tempdir Standard use settings API for location

MagnificentPako:
I suggest that the Tempdir standard moves from a hardcoded location to a path defined by a setting, which can be changed to the user's preferences.

### DIFF
--- a/standards/7-temp-directory.md
+++ b/standards/7-temp-directory.md
@@ -3,38 +3,39 @@
 ## Quick information
 | Information |                                                                 |
 | ----------- | --------------------------------------------------------------- |
-| Version     | 1.0.0                                                           |
+| Version     | 2.0.0                                                           |
 | Type        | Path / Directory                                                |
 | MIME        | `application/directory`                                         |
-| Location    | Temporary Directory is located in `/tmp`                        |
+| Location    | Temporary Directory is located in `settings.get("tempdir")`     |
 
 ### Technical Details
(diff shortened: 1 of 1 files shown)


standards/7-temp-directory.md
```diff
@@ -46,8 +47,14 @@ end
 ```lua
 -- This is a startup file. Append this code to the top of your startup file (or close to it) to adhere to the standard.
 
-fs.delete("/tmp")
-fs.makeDir("/tmp")
+-- check if tempdir setting exists; If it doesn't, set it
+if settings.get("tempdir") == nil then
```
SquidDev: It might be better to do `settings.get("tempdir", "/tmp")` rather than setting it.

SquidDev:
I think this probably needs a little bit more rewording than just replacing every instance of `/tmp` with `settings.get("tempdir", "/tmp")`.

I'm also curious of your use-case here? It seems to add a lot of complications without much tangible benefit.

oeed:
Could you please elaborate on the benefits/use cases of this?

I can see this causing a large number of difficulties and complications. With this you would basically have to include a settings API with every program that wants to save temporary files. We don't have anything else in this API you mention either, so I doubt it would be used that much. The way I see it the more simple this is the more likely it is going to be used.

So unfortunately it's probably a no from me, but if you do have solutions to these issues by all means point them out.

Issue #69, opened 2020-05-11
NFT could be improved by adding support for transparency


Marco4413:
I've been thinking for the last few days that it would be cool if you could change the background/foreground color for pixels on nft images, just like nfp does you could set the color of the background/foreground to a space character to make it transparent.

I was also thinking that if transparency would be added the color of the foreground (if transparent) could be the same as the background of the pixel underneath (in case there was a screen buffer that would support it else the color of the terminal's background).

viluon:
nice

exerro:
nice

PR #70, opened 2020-05-16
Adding transparent colours on nft format

Marco4413:
Adding what was mentioned on issue #69

### DIFF
--- a/standards/6-nft.md
+++ b/standards/6-nft.md
@@ -3,7 +3,7 @@
 ## Quick information
 | Information |                           |
 | ----------- | ------------------------- |
-| Version     | 1.0.0                     |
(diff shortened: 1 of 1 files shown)


oeed:
Hmmm the main issue I have really is that NFT is something that's already pretty widely used and wasn't even created by any of us; so nothing would really change.

Worth noting though that (if memory serves correct) you can just use a space character for transparent foreground.

Marco4413:
You're right but having transparent background and using a space character would mean "don't modify that pixel" and that's what I need. Now I've got a question, what if the color specified isn't a valid hex... Should the color be set to transparent, should the program error or should it use the last color that was set to text/background?

